PR #4, opened 2025-11-07
Motor Driver PCB Design 1.0.0

Scicada46:


bsikar:
1. Make sure to add color to the nets, this was show in the video

bsikar:
2. Make sure the schematic ERC passes

image

bsikar:
3. There are issues on the schematic, when you fix the ERC most of these will be fixed too:

image

3.1: The SPI_CIPO and SPI_CDPI sheet labels are still here, remove them
3.2-3.9: The wire is left floating, make sure that the next label is the last thing on the wire
3.10: The power flag symbol is overlapping the text

bsikar:
4. The ground plane should be selected for all planes
image

bsikar:
5. This is the main source of your current in this, you need to have very wide pours, doing `Draw Filled Zones` and putting a large pour for it, is best.

image
image

bsikar:
6. There are issues with the silk screen and positioning of headers
image

6.1: This is currently on `User.Comments`, but it should be on `F.Silkscreen`
6.2-6.6: The Jx silk screen is far away from the actual headers and in some cases its hard to know which silkscreens go to which header
6.7-6.8: These headers spacing and alignment is wrong, they should have even spacing and be aligned

bsikar:
7. There are poorly made connections
image

7.1: I changed my mind, ignore this one
7.2: The spacing between these 2 lines are too close
7.3: The spacing between these lines are too close and they are at a right angle (90deg) which is very bad, must fix–especially since they will be carrying high current.
7.4: This is at a right angle which is bad
7.5: See the additional two images below, but for these: the line is too close to the other lines and is at 90deg
7.6: The bottom right of this marker has a line thats too close to 90deg
Additional photos:
image
image

bsikar:
8. There are some poorly made connections
image
8.1: This is at a weird angle, make it better
8.2: You could prob use a via so its not wrapping around

bsikar:
9. the DRC says that pin 8 and 21 are not connected, connect them
image

bsikar:
10. I think a power plane would be best. It should be:

Signal
gnd/signal
Power
gnd/signal

bsikar:
Fix the color of these 5V and 6V Bucks
image

bsikar:
Fix the little unconnection end symbols 

image

bsikar:
Our new encoders dont need 5V, they can use 3V3


Remove 74LVC125APW_118

image

bsikar:
Follow this pin out

image

bsikar:
Make sure the headers are updated to this `JST 6-Pin cable`

image

bsikar:
Rename these to be prefixed with MCU_

image

bsikar:
I dont think we need all of these headers, each pin of the other headers like the 6 pin and 8 pin headers can do 0.5A, you need 1VCC and 1GND per each. Do calculations and make sure we are good with those. Also verify with using the JST header (you will need to add this)

image

bsikar:
I dont think this should show the MOTORx_TERMINALx, this needs to go to the MCU so prefix with MCU_ for example MCU_ENCODER1_OUTPUT1

image

bsikar:
Fix the overlap

image

Scicada46:
> Fix the overlap
> 
> image

Fixed the over lap of P6V0_Buck, removed P5V0_Buck due to using 3V0 for motors and no 74LVC125APW_118.

Scicada46:
> Our new encoders dont need 5V, they can use 3V3
> 
> Remove 74LVC125APW_118
> 
> image

Removed

Scicada46:
> Fix the little unconnection end symbols
> 
> image

Fixed

Scicada46:
> I dont think this should show the MOTORx_TERMINALx, this needs to go to the MCU so prefix with MCU_ for example MCU_ENCODER1_OUTPUT1
> 
> image

Added Prefix "MCU_" to encoders

Scicada46:
> Rename these to be prefixed with MCU_
> 
> image

Added Prefix "MCU_" to SPI

Scicada46:
> Follow this pin out
> 
> image

Added connectors for motors

bsikar:
Recived 
[Motor driver board V1.1.0 (1).pdf](https://github.com/user-attachments/files/23616998/Motor.driver.board.V1.1.0.1.pdf)

Reviewed
[Motor driver board V1.1.0.pdf](https://github.com/user-attachments/files/23617005/Motor.driver.board.V1.1.0.pdf)

bsikar:
#8

PR #8, opened 2025-11-20
Motor Driver Board (V1.1.1)

Scicada46:


bsikar:
1. fix the silk screen issue, yours is on user comments
2. make the 6V power plane power better
3. select the F.Cu copper as well for the gnd pour
4. fix the via sizes, make them normal
5. make sure your trace widths are the correct size: https://www.digikey.com/en/resources/conversion-calculators/conversion-calculator-pcb-trace-width?msockid=0c6b153eb2306dd231fd03b2b3bb6c3f
6. add mounting holes following this video: https://youtu.be/Pvz9BWNdG44?si=tDYaUPVq0gEuILN1&t=4556
7. for those the mouting holes use: 
image
8. fix the table in the schematic to reference the motor pinout, verify with the official docs dont just look at ur circuit
9. move the 220n and 100n caps close to the ic

bsikar:
@Scicada46 I am still waiting on this

bsikar:
(1)
All silk screen needs to be:
Width: 1mm
Height: 1mm
Thickness: 0.15mm

(2)
The mounting holes are not places properly, they are also not shown on the schematic. Both needs to be fixed.

PR #11, opened 2025-11-26
Fuel Gauge Version 1.0.0

Scicada46:


bsikar:
(1)
All silk screen needs to be:
Width: 1mm
Height: 1mm
Thickness: 0.15mm

(2)
The mounting holes are not places properly, they are also not shown on the schematic. Both needs to be fixed.

(3)
This is a 4 layer board, but can easily be made into a 2 layer, also the physical size of the board seems awfully large

(4)
LED placement seems very far away

PR #20, opened 2025-12-16
Jbartz/afe

Jhews6:


Jhews6:
image


Add the Positive and negative Markers, so we dont screw up wiring

Jhews6:
image
Change all silk screens to have these dimensions

PR #26, opened 2025-12-17
Add battery management, configuration, and firmware update proto services

CesMag:
## Summary

This PR adds three new protobuf services to expose firmware capabilities for RPi5↔ESP32 communication:

### Battery Management Service (`battery_management.proto`)
- BQ7850 BMS state monitoring (cells, temps, current, SOC, status)
- Protection threshold configuration (OV/UV/OC/OT limits)
- Cell balancing control and status
- FET control for charge/discharge
- Device info and reset commands

### Configuration Service (`configuration.proto`)
- Runtime NVS configuration management
- Motor PID gains (4 motors, reuses `PidConfig` from motor_control.proto)
- Current sensor and temperature calibration
- Safety thresholds (overcurrent, thermal shutdown, cell imbalance)
- Timing parameters (control loop, telemetry, comm timeout)
- Validation with detailed error reporting

### Firmware Update Service (`firmware_update.proto`)
- OTA chunked firmware transfer (~1KB chunks)
- Progress tracking with ETA and transfer speed
- Validation and rollback support
- Reboot and mark-valid commands
- Detailed error handling (12 specific error codes)

### Supporting Files
- **nanopb options** for ESP32 static allocation:
  - `battery_management.options`: max 16 cells, 3 temp sensors, 32-char strings
  - `configuration.options`: max 4 motors, 8 validation errors
  - `firmware_update.options`: 1KB chunk data, 32-char version strings

### Test Coverage
Comprehensive table-driven Kotlin tests (JUnit 5 parameterized):
- `BatteryManagementTest.kt`: Cell voltages, temperatures, SOC, safety faults, device info
- `ConfigurationTest.kt`: PID gains, calibration, thresholds, validation status
- `FirmwareUpdateTest.kt`: Update states, chunk streaming, progress tracking

## Unit Conventions
Following Boston Dynamics style guide:
- Voltage: `_mv` (millivolts)
- Current: `_ma` (milliamps)
- Temperature: `_deci_celsius` (0.1°C units from BQ7850)
- Power: `_mw` (milliwatts)
- Time: `_ms` for config, `_us` for timestamps

## Test plan
- [x] `buf lint` passes
- [x] `buf format` applied
- [x] `buf generate` produces Kotlin/TypeScript/nanopb code
- [x] All Kotlin tests pass
- [ ] CI workflow validates changes


star-proto/tests/kotlin/src/test/kotlin/com/star/proto/BatteryManagementTest.kt
```diff
@@ -0,0 +1,864 @@
+// BatteryManagementTest.kt - Battery Management Protobuf Serialization Tests
```
bsikar: The full file is wrong, other kotlin is wrong too.

  1. Remove all data classes - Delete CellVoltageTestCase, TemperatureTestCase, CurrentTestCase, SocTestCase, and ProtectionTestCase. These add unnecessary abstraction.
  2. Delete the entire companion object - Remove all @MethodSource provider methods (cellVoltageTestCases(), temperatureTestCases(), etc.). They make test data hard to find and understand.
  3. Replace @MethodSource with @CsvSource - Convert all parameterized tests to use inline @CsvSource annotations instead of companion object methods. Put test data directly in the annotation above each test.
  4. Simplify test method signatures - Instead of taking a single data class parameter, use explicit parameters with descriptive names (e.g., v1: Int, v2: Int, v3: Int, v4: Int, packMv: Int).
  5. Inline test data - Move all test case data from the companion object directly into @CsvSource annotations. Each row of CSV should be one test case with a description as the last parameter.
  6. Use backtick test names - Follow Kotlin convention: fun `cell voltage data serializes correctly for 4S pack`()
  7. Keep it simple - Remove complex table-driven patterns. If someone needs to understand a test, they should see the test data right there in the test method, not hunt through a companion object.
  9. Remove unnecessary comments - Keep only the file header. Test names should be self-documenting.
  10. you need to move all the data classes into the actual source, they shouldnt be in the tests:
  11. never nest test classes, each test file needs to correspond to a actual class

PR #35, opened 2025-12-30
Remove BQ25798 buck boost charger driver

CesMag:
## Summary
- Remove BQ25798 buck boost charger driver (no longer used in hardware)
- Update BMS config to use BQ4050RSMR fuel gauge on SMBUS
- Keep generic rx_fuel_gauge SMBUS driver (compatible with BQ4050)

## Changes
- Delete `rx_bq25798.c` and `rx_bq25798.h`
- Update `CMakeLists.txt` build configuration
- Update `hardware_config.h` BMS pin assignments

## Hardware Context
Replaced the buck boost charger IC with BQ4050RSMR BMS IC for battery management via SMB bus.


bsikar (changes requested):
This is very generic, we need to have an actual implementation for this new fuel gauge

PR #248, opened 2026-02-03
TOM_MCU PCB

Scicada46:
I think that we can remove Priority Switchover unless we wanted to add the 3v3 buck test pin into it. Or remove it and have BUS_3V3 be the board 3v3. Plus I left the other pins currently disconnected on the R5F5, but left the labels there currently, for if we wanted to add more test points.


Jhews6:
image

image


These are filtering cap near mcu need to be closer

Jhews6:
image

image

Place as close to mcu pin as possible here asweell also filtering

Jhews6:
image

We are doing high speed comm lines, we need a uninterrupted ground layer to separate high speed signals and noisy signals. Usually layer 2 is a sold ground layer

Jhews6:
image

on usbc you have to do something called a differential pair. You want the the lengths of the traces to match up almost identically

Jhews6:
image
same as this line the p and n of the trace should be the same length, in addition to this it needs to be as close as possible to the mcu, try to keep on one layer. These comm lines are very important and very sensitive and will fuck everything up if they are wrong, most important part of pcb

This video will help you with the pcb please watch:
https://www.youtube.com/watch?v=Itsrdc8tX7M

Jhews6:
image
add room in between components, will be hard to solder

Jhews6:
image
same here room for soldering

Jhews6:
image
Keep this on one layer this is a sensitive clk line

Jhews6:
Overall try sticking to layer 1 and 4 for signal lines, keep layer 2 gnd. Layer 1 is usually for high speed, sensitive lines, i2c, usbc, etc. i2c is alright to go down a layer just avoid if possible if no choice dont worry tm.

Jhews6:
image
you have traces that go no where put no connects here

Jhews6:
image
i got different values from you on ur usb lines

Jhews6:
image
move edge closer to usbc both of them those shits are always hard to plug in

CesMag (changes requested):
Still pending

PR #252, opened 2026-02-05
Review Completed

jared-bartz-tamu:
Main thoughts based off of looking between provided table and board

Everything looks fine for the most part, I checked using the table in the meeting notes, all the issues I found relate to the PMOD pins, with everything else going to the right spot.

ALL PINS LISTED BELOW ARE PMOD PINS THAT GO TO NOTHING:
Pins 27-28
Pins 41-44
Pins 47-50
Pins 55-58

We should probably just X these unless we are using them. I think it would also be nice if we renamed the LED and Sonar pins that come from the PMOD_GPIO pins to something else. 
image
I think these could have their lines renamed to fit with the rest. Maybe SONAR_TRIG_# and SONAR_ECHO_# for the sonar pins, and just LED_# for the LEDS.

Otherwise, I didn't notice anything else. I went through one by one and traced the lines, they all seemed to go to the right place.

### DIFF
--- a/schematic/#auto_saved_files#
+++ b/schematic/#auto_saved_files#
@@ -1,1 +1,0 @@
-C:\Users\redwi\Documents\GitHub\STAR\schematic\_autosave-STAR_MCU.kicad_sch
--- a/schematic/#auto_saved_files#
+++ b/schematic/#auto_saved_files#
@@ -1,0 +1,1 @@
+C:\Users\redwi\Documents\GitHub\STAR\schematic\_autosave-STAR_MCU.kicad_sch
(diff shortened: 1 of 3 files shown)


bsikar (changes requested):
image

This is not connected



image

These labels existed for PMOD use, remove all of the use for things with RX72N_PIN_xxxxxxx

image


Comments lead to confusion, remove them -- do the same for all weird random comments that you put during your review

CesMag:
@jared-bartz-tamu have you worked on accepting brightons reviews?

PR #257, opened 2026-02-07
Power Board Review.

jared-bartz-tamu:
To be honest, I've been looking and genuinely can't find anything wrong outside of nitpicks. I'm not an expert on power boards, but i was able to follow the traces and both the schematic and pcb made sense to me.  All I did was move the MOSFET part names on the schematic since they were overlapping with the lines.

Small note, when I made this branch, I noticed it said some edits might not have been saved, and also failed to load an auto saved file. I'm going to check in with Jeremie tomorrow and make sure everything is fine on that front. If it's not, I'll quickly redo the review.

image

image

This is pretty much all I did. Sorry it isn't more but I'm not sure what else to add when I can't find any more issues, lol.

### DIFF
--- a/schematic/#auto_saved_files#
+++ b/schematic/#auto_saved_files#
@@ -1,1 +1,0 @@
-C:\Users\jerem\Project Star\STAR\schematic\_autosave-STAR_POWER_BOARD.kicad_sch
--- a/schematic/STAR_POWER_BOARD.kicad_sch
+++ b/schematic/STAR_POWER_BOARD.kicad_sch
@@ -12364,7 +12364,7 @@
 			)
 		)
 		(property "Value" "CSD17577Q3A"
-			(at 152.146 224.79 0)
+			(at 144.272 225.044 0)
 			(effects
 				(font
 					(size 1.27 1.27)
@@ -26933,7 +26933,7 @@
 			)
 		)
 		(property "Value" "SSM6J507NU-LF"
-			(at 221.996 85.598 0)
+			(at 214.63 74.676 0)
 			(effects
 				(font
 					(size 1.27 1.27)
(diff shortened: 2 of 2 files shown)


jared-bartz-tamu:
Adding to this, here are some things ive seen that probably arent issues, but just want to make sure

image

Is this island here supposed to be intentional? The GND vias here make me think it is, so it's likely fine.

image

This turn feels awfully tight, but honestly it's also fine. A lot of my reviews are going to be small nitpicks since the big picture looks great.

image

I think this could be improved? Not really sure, but I think you could more easily just make it a direct diagonal line.

image

These two could probably be pushed more forward to look nicer? I notice they all make great contact with the poly, so it likely isnt an issue outside of aesthetics.

image

I think this U3 could be moved to be closer to the actual component.

image

J7 and J6 need their silks moved to be actually visible (unless that's intentional), and I think J2 is somehow lacking any silkscreen text entirely.

image
 
more of a nitpick than anything, but i think it'd look nicer if the Q2 and Q3 text were moved to be to the sides of the MOSFETs. Theyre a bit slanted as is and it looks awkward.

image

The traces are going through the U1 text. Maybe move it to the side?

image

same with r11, could probably move the silk under.

image

once again this island feels very intentional, but I feel the need to put it here just to make sure, since i know the back plate is a massive ground pour.

image

The GND plate for C32 is not fully touching the GND pour. Likely fine but still, bringing everything I notice to attention.

Aside from all these little things, I didn't seen any major things like bottlenecks or places that didn't connect. Also, I'd recommend **NOT** merging this into the main branch, as this pull request was made before you added the screw. Just look at what I pointed out and see if that's intentional or not.

PR #264, opened 2026-02-10
Jhews/power board no6 v

Jhews6:
new power board with no 6v buck since it is now external


bsikar:
image

bsikar (changes requested):
1) Missing test points for the input and output of this
2) Missing silk screen for input and outputs for this
3) Pours could be filled in more

bsikar:
image


4) Put `x` and remove the extra talking about Jx/Jx/Jx bc idk why thats there
5) rename this to say its M3, **make sure its M3** or whatever we said it should be for micol

PR #280, opened 2026-02-12
144 Pin MCU Review

jared-bartz-tamu:
For Brighton to look at, it's our new pin package.


bsikar (changes requested):
- Remvove the headers going to the mcu the 15ish pin headers---i dont remember the exact 
- Finish BOM

PR #349, opened 2026-02-18
fix(gateway): auto-respond RESET_ACK to received RESET frames

CesMag:
## Summary
- `dispatchControlFrame()` was missing a `case frame.FrameTypeReset:` branch, causing firmware-initiated RESET frames to silently fall through to `default` and be discarded without any response
- Adds the missing case and a new `sendResetAck()` method, satisfying the requirement in `TRANSPORT_ARCHITECTURE.md` §Control Frame Dispatching: *"Both firmware (C) and gateway (Go) auto-respond RESET_ACK to RESET"*
- `sendResetAck()` is modelled directly after `sendPong()` — it bypasses the operations gate to avoid deadlock and sends `FrameTypeResetAck` (0xFE) with no payload, as specified

## Test plan
- [x] `go build ./...` — zero errors
- [x] `go test ./...` — all 13 packages pass
- [x] Code review checklist:
  - [x] NASA Power of 10 rules followed (null-check on active transport, error checked and logged)
  - [x] SOLID principles applied (new method mirrors existing `sendPong` pattern)
  - [x] No magic numbers
  - [x] Inclusive terminology used

## Summary by CodeRabbit

* **Bug Fixes**
  * RESET control frames are now explicitly acknowledged and consumed, reliably resetting session state while preserving heartbeat updates and avoiding control-path deadlocks.
  * Control-frame send timing unified to a short, non-blocking deadline to improve reliability of control replies.
* **Tests**
  * Added a test validating RESET acknowledgement and session reset behavior.

### DIFF
--- a/star-gateway/internal/manager/heartbeat.go
+++ b/star-gateway/internal/manager/heartbeat.go
@@ -28,8 +28,9 @@ const (
 	// pingPayloadSize is the size of PING frame payload (4-byte counter).
 	pingPayloadSize = 4
 
-	// pingSendTimeout is the timeout for sending a PING frame.
-	pingSendTimeout = 100 * time.Millisecond
+	// controlFrameSendTimeout is the deadline for sending any control frame reply
+	// (PONG, RESET_ACK, etc.). Short enough to be non-blocking on the receive path.
+	controlFrameSendTimeout = 100 * time.Millisecond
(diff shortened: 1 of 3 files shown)


CesMag:
@coderabbitai review

PR #364, opened 2026-02-22
Bsikar/mcu

Jhews6:
Please check for high-current paths and make sure they use sufficiently thick traces or polygons with multiple connections. Check for any 90-degree angles, and review the DRC. There are many issues I ignored because they are footprint-related, but I need someone to double-check that I am not missing anything major. Verify that my screw terminals are facing the correct direction. Make sure there are no copper islands. Additionally, ensure that all high-power paths are not going through bottlenecks and double-check for any narrow sections. Take your time I need a very in depth review, this is our only PCB so we can not mess this up.


Scicada46:
Screenshot 2026-02-22 at 5 30 45 PM
Vias outside of edge cuts for Pack+

Scicada46:
near to 90 degree bend 
Screenshot 2026-02-22 at 5 35 39 PM

jared-bartz-tamu:
Ill be commenting as i go, then try and organize this later. Screw terminals are facing the right directions.

### 90 DEGREE/NEAR 90s

image

Near 90 degree turn by U3 and J7 

image

another near 90 at U10

image

another near 90 coming off U5 but i dont see a better way to do this, should be fine

image

another near 90 

image

not necessarily near 90, but still a really tight turn that can be smoothed out

image

90 degree turn from the via

image

This feels pretty tight for such a big trace, maybe this works better?

image

another 90 from the via

image

90 from via

### DRC ISSUES:

image
Your ENC2_TCKLA line runs through multiple of the switch out vias

Honestly the rest of the DRC issues are weird footprint things. This is the only glaringly wrong issue I see

### BOTTLENECKS 

Going to be honest, these might be stretches, but better to point out than not.

image

This /SCI_USBC_5V0 pour is pretty rough looking. I think the pour is still large enough to not burn anything, but if its possible to nudge the PS3 component slightly to the left, it might reduce any risks.

image
This AVCC0 pour gets squeezed pretty tight here to make it to the via, then 180s into the component. You said ICs are ok so im guessing this is nothing, but it's a very small trace for such a big pour. I think the surrounding pours could be shrunk slightly without compromising them in order to increase the size of this one.


### NITPICKS (?)

Areas where i think we can reduce angle

image

the via could be shifted up and left to make this smoother?

image

could be moved up

image

This could be moved over here to reduce the angle

PR #381, opened 2026-02-25
Jbartz/dma crc support

jared-bartz-tamu:
It should be fully operational. I'd say run more tests on it to see how it does, but from everything I've done it seems to be good. Phase 1 and 3 should be fully implemented, with bits of phase 2? I think the only things left should be Doxygen documentation since I wasnt sure how to do that, and maybe making the API prettier. Sorry this took so long, It took a long time to read through and understand the code and what bits i needed to flip and alter, and I've had other schoolwork keeping me busy. This was definitely a challenge to remember how to read the code and get it logically sound, but it's def helped get the rust off.

## Summary by CodeRabbit

* **New Features**
  * Added a DMA controller driver supporting 8 channels, safe blocking transfers, timeouts, and a simple blocking-transfer API.
  * Exposed configurable data-size, transfer-mode, and address-mode options with robust input validation and bounded transfer limits.
  * Integrated optional DMA acceleration for CRC32 with a configurable threshold and automatic CPU fallback for smaller or unsupported transfers.


e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
+/* libs/rx_crc/inc/rx_dmaca.h */
+
+/**
+ * @file    rx_dmaca.h
+ * @brief   DMACA driver for RX72N
```
bsikar: remove the spacing here, reference the other files on how this should look

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+ *
+ * @par SOLID Principles
+ * - **Single Responsibility:** DMACA channel management and transfer control only
+ * - **Dependency Inversion:** Abstracts error codes via rx_err_t enum
+ *
+ * @see rx_dmaca.tex
+ * @author STAR Team
+ * @date 2026-02-22
+ * @version 1.0.0
+ * @copyright Copyright (c) 2026 STAR Project - MIT License
```
bsikar: no need to specify the MIT license, reference other files for exact way on how we do it

bsikar: if other files show with the `- MIT` thats wrong and we need to make a ticket for it

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+ */
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+#include "rx_err.h"   /* rx_err_t, k_rx_ok, k_rx_err_timeout, k_rx_err_nack */
```
bsikar: remove the comment

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+
+/* =========================================================================
+ * Public constants
+ * ========================================================================= */
```
bsikar: line shouldnt end with */ the followign line should

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+ * // Example: iterate over all valid DMACA channels
+ * for (uint8_t ch = 0; ch < k_dmaca_num_channels; ch++) {
+ *     rx_dmaca_abort(ch);
+ * }
+ * @endcode
+ *
+ * @see rx_dmaca_configure, dmaca_config_t, validate_channel
+ * @since 1.0.0
+ */
+typedef enum : uint32_t {
```
bsikar: why uint32_t if the max value is a 10-bit?

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+ * @endcode
+ * @note All values are compile-time constants and may be used in static
+ *       initializers or case statements.
+ *
+ * @see rx_dmaca_crcdir_reg_ptr, dmaca_config_t
+ *
+ * @since 1.0.0
+ */
+typedef enum : uintptr_t {
+    k_dmaca_crcdir_phys_addr = 0x00088284U, /**< CRCDIR register */
```
bsikar: is this 4 spaces instead of 2? did you not run the formatter script?

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
+/* libs/rx_crc/inc/rx_dmaca.h */
```
bsikar: This file belongs in libs/rx_hal/inc/, not libs/rx_crc/inc/. The DMAC is a general-purpose hardware peripheral - it has nothing specific to CRC. Placing it here means any future consumer of DMA (SPI, UART, etc.) would have to import the CRC library to get it. Move to libs/rx_hal/ alongside rx72n_dmac_regs.h, rx72n_crc_regs.h, etc.

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+ * for (uint8_t ch = 0; ch < k_dmaca_num_channels; ch++) {
+ *     rx_dmaca_abort(ch);
+ * }
+ * @endcode
+ *
+ * @see rx_dmaca_configure, dmaca_config_t, validate_channel
+ * @since 1.0.0
+ */
+typedef enum : uint32_t {
+    k_dmaca_max_transfer_count = 1024U, /**< 10-bit DMTC field max (1..1024) */
```
bsikar: k_dmaca_num_channels = 8U already exists in rx72n_dmac_regs.h as k_dmac_channel_count = 8U (line 173). Don't duplicate it here - use the existing constant.

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
```diff
@@ -0,0 +1,587 @@
...
+ *   volatile uint32_t *p_crcdir = (volatile uint32_t *)k_dmaca_crcdir_phys_addr;
+ * @endcode
+ * @note All values are compile-time constants and may be used in static
+ *       initializers or case statements.
+ *
+ * @see rx_dmaca_crcdir_reg_ptr, dmaca_config_t
+ *
+ * @since 1.0.0
+ */
+typedef enum : uintptr_t {
```
bsikar: Address enum underlying type should be uint32_t, not uintptr_t. All hardware address enums in this project use uint32_t - see rx72n_dmac_regs.h, rx72n_crc_regs.h, rx72n_cmt_regs.h. The RX72N is a 32-bit platform so uint32_t is sufficient and consistent.

e2-studio-star-rx72n-firmware/libs/rx_crc/inc/rx_dmaca.h
bsikar: k_dmaca_crcdir_phys_addr = 0x00088284U is a duplicate magic address. CRCDIR is already accessible as &crc_regs()->crcdir via rx72n_crc_regs.h. Hardcoding the raw offset here will silently diverge if the CRC register layout ever changes. Remove this enum entirely - callers should use the register accessor.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * volatile rx_dmac_channel_regs_t *p_ch = (volatile rx_dmac_channel_regs_t *)channel_addr;
+ * @endcode
+ *
+ * @see dmac_ch, k_dmac0_base_addr, k_dmac_channel_size
+ * @since 1.0.0
+ */
+/* Local alias for hardware register layout constant from rx72n_dmac_regs.h */
+
+typedef enum : uint32_t {
+    DMACA_CHANNEL_STRIDE = k_dmac_channel_size,       /**< bytes between channels */
```
bsikar: DMACA_CHANNEL_STRIDE uses SCREAMING_SNAKE_CASE which the project reserves for macros only. Enum constants must use the k_ prefix: k_dmaca_channel_stride. This applies throughout the file - see all the constants below.

e2-studio-star-rx72n-firmware/libs/rx_hal/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * ch->DMSTS = (uint8_t)DMACA_DMSTS_CLEAR;
+ *
+ * // Trigger software request
+ * ch->DMAMD |= (uint8_t)DMACA_SWREQ_REQUEST;
+ * @endcode
+ *
+ * @see DMACA_DMINT_CLEAR, DMACA_DMSTS_CLEAR, DMACA_SWREQ_REQUEST
+ * @since 1.0.0
+ */
+typedef enum : uint8_t {
```
bsikar, line 110: DMACA_DMINT_CLEAR, DMACA_DMSTS_CLEAR, DMACA_DMCRB_CLEAR, DMACA_SWREQ_REQUEST - all must use k_ prefix: k_dmaca_dmint_clear, etc. See CLAUDE.md naming conventions.

e2-studio-star-rx72n-firmware/libs/rx_hal/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * empty bitmap and the base bit value used when shifting by a channel index.
+ * Using typed enum values avoids magic literals and clarifies intent when
+ * manipulating the bitmap.
+ *
+ * @invariant Only the two enumerators are used; shifting by channel index is
+ *            performed with DMACA_CHANNEL_BITMAP_BIT.
+ *
+ * @since 1.0.0
+ */
+typedef enum : uint8_t {
```
bsikar, line 133: DMACA_CHANNEL_BITMAP_NONE, DMACA_CHANNEL_BITMAP_BIT - same, must be k_dmaca_channel_bitmap_none, k_dmaca_channel_bitmap_bit.

e2-studio-star-rx72n-firmware/libs/rx_hal/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * // Example: enable DMA transfer engine for a channel
+ * ch->dmcnt |= (uint8_t)DMACA_DTE_ENABLE;  // Set DTE = 1
+ * // Example: disable DMA transfer engine
+ * ch->dmcnt &= ~(uint8_t)DMACA_DTE_ENABLE;  // Clear DTE bit to disable
+ * @endcode
+ *
+ * @see DMACA_DTE_ENABLE, DMACA_DTE_DISABLE, rx_dmaca_start, rx_dmaca_abort
+ * @since 1.0.0
+ */
+typedef enum : uint8_t {
```
bsikar, line 161: DMACA_DTE_DISABLE, DMACA_DTE_ENABLE - must be k_dmaca_dte_disable, k_dmaca_dte_enable.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * uint32_t timeout = (transfer_size * K_DMACA_CYCLES_PER_XFER * 
+ *                      K_DMACA_MARGIN_NUMERATOR) / K_DMACA_MARGIN_DENOMINATOR;
+ * rx_dmaca_wait(channel, timeout);
+ * @endcode
+ *
+ * @see K_DMACA_CYCLES_PER_XFER, K_DMACA_MARGIN_NUMERATOR, K_DMACA_MARGIN_DENOMINATOR, 
+ *      rx_dmaca_transfer_blocking, rx_dmaca_wait
+ * @since 1.0.0
+ */
+typedef enum : uint32_t {
```
bsikar: K_DMACA_CYCLES_PER_XFER, K_DMACA_MARGIN_NUMERATOR, K_DMACA_MARGIN_DENOMINATOR - same issue. These have an uppercase K_ prefix which is even more wrong. Must be k_dmaca_cycles_per_xfer, etc.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * // Example: construct DMTMD value with normal mode and byte size
+ * uint16_t dmtmd = ((DMTMD_MD_NORMAL << DMTMD_MD_SHIFT) | 
+ *                    (DMTMD_SZ_BYTE << DMTMD_SZ_SHIFT));
+ * ch->dmtmd = dmtmd;
+ * @endcode
+ *
+ * @see dmtmd_mode_t, dmtmd_size_t, rx_dmaca_configure, DMTMD_MD_SHIFT, DMTMD_SZ_SHIFT
+ * @since 1.0.0
+ */
+typedef enum : uint16_t {
```
bsikar: DMTMD_MD_SHIFT, DMTMD_SZ_SHIFT, DMTMD_MD_NORMAL, DMTMD_MD_BLOCK, DMTMD_MD_REPEAT, DMTMD_SZ_BYTE, DMTMD_SZ_LONGWORD - all must use k_ prefix.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * // Example: construct DMAMD register with increment source, fixed destination
+ * uint16_t dmamd = ((DMAMD_SM_INCREMENT << DMAMD_SM_SHIFT) |
+ *                    (DMAMD_DM_FIXED << DMAMD_DM_SHIFT));
+ * ch->dmamd = dmamd;
+ * @endcode
+ *
+ * @see dmamd_source_mode_t, dmamd_dest_mode_t, DMAMD_SM_SHIFT, DMAMD_DM_SHIFT
+ * @since 1.0.0
+ */
+typedef enum : uint16_t {
```
bsikar: DMAMD_SM_SHIFT, DMAMD_DM_SHIFT, DMAMD_SM_FIXED, DMAMD_SM_INCREMENT, DMAMD_SM_DECREMENT, DMAMD_DM_FIXED - all must use k_ prefix.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ *
+ * @note Thread safety: safe to call from any context; the calculation does not
+ *       modify shared state.  The returned pointer should only be used with
+ *       proper synchronization if the channel is accessed concurrently.
+ *
+ * @see rx_dmaca_start, rx_dmaca_abort, k_dmac0_base_addr, DMACA_CHANNEL_STRIDE
+ * @since 1.0.0
+ */
+ 
+static volatile rx_dmac_channel_regs_t * dmac_ch(uint8_t p_ch)
```
bsikar: dmac_ch() should be named internal_dmac_ch() - project style requires internal_ prefix for static functions (see CLAUDE.md). More importantly, this function reimplements dmac_channel() which already exists in rx72n_dmac_regs.h:637. Delete this function and call dmac_channel() directly instead.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ *
+ * @note This is an internal, fast helper; callers should still handle the
+ *       error return value appropriately.  Uses RX_ASSERT macros to validate
+ *       preconditions during debug builds.
+ *
+ * @since 1.0.0
+ *
+ * @see k_dmaca_num_channels, k_rx_err_nack, k_rx_ok, RX_ASSERT, validate_channel
+ */
+static rx_err_t validate_channel(uint8_t channel)
```
bsikar: validate_channel() must be internal_validate_channel() per the internal_ prefix rule for static functions.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -168,25 +168,33 @@
...
- * - **RESOLUTION:** Each call sets DORCLR first, ensuring clean state
- *
- * **Conclusion:** Safe for concurrent use without mutex (verified by testing).
+ * - **Concurrent access (CPU-only):** Multiple threads can compute CRCs
+ *   simultaneously without serialization.
+ * - **Concurrent access (DMA path):** The DMA engine and CRC hardware must
+ *   be accessed exclusively.  A module-level lock (`crc_lock_acquire()`/
+ *   `crc_lock_release()`) is taken around the entire transaction when
+ *   `len >= RX_CRC32_DMA_THRESHOLD`.  This prevents interleaving of DMA
+ *   transfers and remainder byte writes that would otherwise corrupt results.
```
bsikar: This analysis is wrong. The "RESOLUTION" listed is not actually a resolution - it describes the corruption. Thread A's DORCLR resets the accumulator, then Thread B's DORCLR resets it again before Thread A finishes writing. Thread A then writes its bytes into a freshly-reset accumulator that Thread B may also be writing to simultaneously. One CRCDIR register exists in hardware. There is no isolation between callers. The CPU-only path is not thread-safe. Remove the thread-safe claim or add a mutex covering DORCLR through CRCDOR read.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
bsikar, line 614: crc_regs() == nullptr - nullptr is C++ syntax. C uses NULL.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -576,6 +786,93 @@ rx_err_t rx_crc_deinit(void)
   return k_rx_ok;
 }
 
+/* =============================================================================
+ * DMA helper (only compiled when RX_CRC32_USE_DMA is defined)
```
bsikar, line 870: RX_CRC32_USE_DMA is never defined in any CMakeLists.txt. All the DMA code in rx_crc32_hw.c and the entire rx_dmaca.c driver are compiled into the binary but the DMA path in rx_crc32_ieee_impl() is unreachable. Either wire up the define in CMake with an option, or don't merge the DMA code until it's actually usable.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+     *
+     * Secondary check: DTE auto-clears to 0 after block completion.
+     *
+     * The timeout counter counts loop iterations, not exact CPU cycles,
+     * but each iteration reads a volatile peripheral register (>= 4 cycles
+     * at 60 MHz PCLKB / 240 MHz ICLK), so the bound is conservative.
+     */
+    uint32_t remaining = timeout_cycles;
+
+    while (((ch->dmsts & k_dmsts_act) != 0U) ||
```
bsikar: Loop structure violates NASA Rule 2 spirit
The rule requires fixed, statically provable loop bounds. The project does this with for. The for form makes the bound immediately visible to a code reviewer and static analyzer. The while form hides the bound inside the body and off-by-one is subtle: with timeout_cycles = N you actually get N+1 iterations before timeout.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+           ((ch->dmcnt & (uint8_t)DMACA_DTE_ENABLE) != 0U))
+    {
+        if (remaining == 0U) {
+            /* Timeout: forcibly disable channel before returning. */
+            ch->dmcnt &= ~(uint8_t)DMACA_DTE_ENABLE;
+            s_channel_configured &= (uint8_t)(~(DMACA_CHANNEL_BITMAP_BIT << channel));
+            return k_rx_err_timeout;
+        }
+        remaining--;
```
bsikar: Pure busy-wait burns CPU in a ThreadX system
The loop does nothing between polls - no tx_thread_sleep(0), no yield. A DMA transfer for 1024 bytes takes ~5120 PCLKB cycles. The caller spins the CPU for the full duration. Every ThreadX task with lower or equal priority is starved. Either use a DMA completion interrupt with a semaphore, or yield with tx_thread_sleep(1) between polls.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
bsikar: Two separate volatile reads in while condition. These are two separate MMIO reads. Between them, DMA activity can change. Check dmsts alone — the ACT bit is the hardware's authoritative "transfer in progress" indicator. Checking DTE as well is redundant for block mode.

e2-studio-star-rx72n-firmware/libs/rx_hal/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * @post On k_rx_err_timeout: Transfer engine disabled, configured flag cleared.
+ * @post On k_rx_err_nack: Channel state unchanged.
+ *
+ * @note Implements a bounded polling loop; see NASA Rule 2 for reasoning.
+ *       This function does **not** disable interrupts or yield the CPU.
+ *
+ * @see rx_dmaca_start, rx_dmaca_configure, rx_dmaca_transfer_blocking
+ * @since 1.0.0
+ */
+rx_err_t rx_dmaca_wait(uint8_t channel, uint32_t timeout_cycles)
```
bsikar, line 916: The Doxygen on all three says @pre rx_dmaca_init() has been called, but none of them verify it at runtime. Calling any of these before rx_dmaca_init() accesses hardware registers with DMST=0 (module clock off). Add a RX_VALIDATE_INIT(s_dmaca_initialised, ...) check at the top of each.

e2-studio-star-rx72n-firmware/libs/rx_hal/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ * @post s_dmaca_initialised == true
+ * @post s_channel_configured == 0U
+ *
+ * @note This routine is NOT thread-safe and should be called once from a
+ *       single initialization context prior to any RTOS tasks using DMACA.
+ *
+ * @see dmac_dmast_reg(), k_dmast_dmst
+ * @since 1.0.0
+ */
+rx_err_t rx_dmaca_init(void)
```
bsikar, line 649: The @post says "All 8 channels are disabled (DMCNT.DTE == 0 for channels 0-7)" but the code only does:
```
*dmac_dmast_reg() |= (uint8_t)k_dmast_dmst;  // enable module
s_dmaca_initialised  = true;
s_channel_configured = 0U;
```

It clears the software bitmap but never writes any hardware register to clear DTE per-channel. If DMA was running before init is called, channels remain active.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_dmaca.c
```diff
@@ -0,0 +1,1273 @@
...
+ *       rx_dmaca.c and should only be accessed within this translation
+ *       unit.
+ *
+ * @warning Direct modification of this variable is forbidden outside of
+ *          this module.  Only rx_dmaca_init() or other internal helper
+ *          routines are permitted to change its value.
+ *
+ * @since 1.0.0
+ */
+static bool s_dmaca_initialised = false;
```
bsikar: Variable named s_dmaca_initialised (British spelling)

Every other variable in the codebase uses initialized (American). Should be s_dmaca_initialized for consistency.

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -865,12 +1178,49 @@ uint32_t rx_crc32_ieee_impl(const uint8_t* data, uint32_t len)
      * NASA Rule 2: Loop bounded by k_crc_len_max (compile-time constant)
```
bsikar, line 1261: Inverted loop, The loop iterates up to 65535 with an internal break. Should be for (uint32_t i = 0; i < len; i++). The current form adds a branch in the hot inner loop and misleads about what the actual bound is.

bsikar (changes requested):

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -297,20 +305,24 @@
  *
  * @author STAR Team
  * @date 2026-01-28
- * @copyright Copyright (c) 2026 STAR Project - MIT License
+ * @copyright Copyright (c) 2026 STAR Project. MIT License.
```
bsikar: remove MIT

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -237,7 +245,7 @@
  * | **2. Fixed Loop Bounds** | [PASS] FULL | Loop bounded by k_crc_len_max (65535) |
  * | **3. No Dynamic Memory** | [PASS] FULL | Zero malloc/free, all state is static |
  * | **4. Short Functions** | [PASS] FULL | Longest function: rx_crc32_ieee_impl (54 lines) |
- * | **5. Assertions** | [PASS] FULL | Minimum 2 checks per function (nullptr, range) |
+ * | **5. Assertions** | [PASS] FULL | Minimum 2 checks per function (NULL, range) |
```
bsikar: nullptr not NULL

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -297,20 +305,24 @@
...
 
 #include "rx72n_regs.h"
 #include "rx_check.h"
+#include "rx_crc_internal.h"
 #include "rx_gpio_constants.h"
 #include "rx_register_protection.h"
 
+#ifdef RX_CRC32_USE_DMA
+#include "rx_dmaca.h"
+#include "tx_api.h" /* tx_interrupt_control(), tx_thread_sleep() */
```
bsikar: remove comment

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -336,13 +348,189 @@ typedef enum : uint32_t {
   k_crc_bit_set        = 1, /**< Bit set value for register manipulation */
 } rx_crc_hw_loop_constants_t;
 
+/* =============================================================================
+ * DMA Configuration
+ * =============================================================================
+ */
+
+#ifdef RX_CRC32_USE_DMA
```
bsikar: shouldnt be using a macro, if we want dma we will just use it with dma via a method

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -336,13 +348,189 @@ typedef enum : uint32_t {
...
+
+/**
+ * @brief DMACA channel used for CRC data transfer.
+ *
+ * Override at compile time with -DRX_CRC32_DMA_CHANNEL=<n> if channel 0
+ * is already claimed by another driver. Falls back to
+ * `k_crc_dma_channel_default` defined above.
+ */
+#ifndef RX_CRC32_DMA_CHANNEL
+#define RX_CRC32_DMA_CHANNEL (k_crc_dma_channel_default)
```
bsikar: no macros

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -336,13 +348,189 @@ typedef enum : uint32_t {
...
+ *
+ * Below this threshold the CPU loop is faster because DMA setup overhead
+ * (configure + SWREQ latency ~ 30-50 cycles) exceeds the loop savings.
+ * Measured break-even on RX72N @ 240 MHz is approximately 64 bytes; the
+ * default value is `k_crc_dma_threshold_default`.
+ *
+ * Override with -DRX_CRC32_DMA_THRESHOLD=<n> after profiling workload.
+ */
+#ifndef RX_CRC32_DMA_THRESHOLD
+#define RX_CRC32_DMA_THRESHOLD (k_crc_dma_threshold_default)
```
bsikar: no macros

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -336,13 +348,189 @@ typedef enum : uint32_t {
...
+ *   while already holding the lock will deadlock.
+ * - Error paths release the lock prior to returning to prevent deadlock.
+ *
+ * @since Version 1.2.0
+ */
+static volatile bool s_crc_lock = false;
+
+/* Helper helpers for lock acquire/release.  These are defined only when DMA
+ * is enabled because the lock is only needed in that configuration. */
+static void crc_lock_acquire(void)
```
bsikar: name of methodd is wrong, no doc comments

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -488,7 +676,7 @@ static bool s_crc_initialized = false;
  *   }
  *
  *   // Verify peripheral is responsive
- *   if (crc_regs() == nullptr) {
+ *   if (crc_regs() == NULL) {
```
bsikar: nullptr not null

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -535,6 +723,29 @@ rx_err_t rx_crc_init(void)
...
+#ifdef RX_CRC32_USE_DMA
+  /* -----------------------------------------------------------------------
+   * Initialise DMACA for CRC transfers.
+   *
+   * rx_dmaca_init() sets DMAC.DMAST.BIT.DMST = 1.  
+   * Safe to call here because the CRC module clock is already enabled above.
+   *
+   * NASA Rule 7: return value checked before proceeding.
+   * ----------------------------------------------------------------------- */
+  if (!s_dma_initialized) {
```
bsikar, line 807: ugly af

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -637,7 +936,7 @@ rx_err_t rx_crc_deinit(void)
  * | 4096 bytes | 24480 | 102.0 | **40.2** | Sustained peak |
  *
  * **Breakdown (1024-byte buffer):**
- * - Validation: ~10 cycles (nullptr, range checks)
+ * - Validation: ~10 cycles (NULL, range checks)
```
bsikar: nullptr no null

bsikar (changes requested):

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -1155,7 +1219,10 @@ uint32_t rx_crc32_ieee_impl(const uint8_t* data, uint32_t len)
 
 #ifdef RX_CRC32_USE_DMA
```
bsikar, line 1235: we shouldnnt have RX_CRC32_USE_DMA

bsikar (changes requested):

CesMag (changes requested):
@jared-bartz-tamu Since we are using this for error checking, please make sure the DMA is configured to feed both the incoming communication data and the appended CRC code sequentially into the CRC Data Input Register (CRCDIR). If the data is error-free, the hardware calculation will result in the CRC Data Output Register (CRCDOR) reading exactly `0000 0000h`

bsikar:
@jared-bartz-tamu also for this like what cesar said you will also have to update our code to make sure we are not using hardware crc 8 and crc 32 in our code base, but instead we should be using this DMA. For this though, we need to have both hardware and dma support for these. so once you get to it make sure you update all the tests and the code so it links proerly. if for some reason your dma fails when we test on hardware we can easily use the other version of how the crc is handled. you might want to abstract this out so our libs call this better, reference how we have done our other libs with their structure and includes.

CesMag (changes requested):
Add unit tests, it is best practice to always make sure to have unit tests of the feature you are adding. In this case, it would be useful for you to read the datasheet, and understand how DMA and CRC work and then have I would say around 75% test coverage for this specific PR.

You can take a look at existing unit tests over at: `e2-studio-star-rx72n-firmware/tests` to see how they work, make sure to link things properly.

bsikar:
Hey, quick one — `RX_CRC32_DMA_CHANNEL` and `RX_CRC32_DMA_THRESHOLD` are 
used in three places in `rx_crc32_hw.c` but I can't find them defined 
anywhere in the codebase. I grepped the whole tree. CMakeLists also turns 
DMA ON by default, so this isn't a dead code path — the default build 
just... doesn't compile.

Also per CLAUDE.md, integer constants need to be C23 typed enums, not 
preprocessor macros. So even once you define them, the form needs to change:

  typedef enum : uint8_t  { k_crc_dma_channel   = 0U;  } crc_dma_channel_t;
  typedef enum : uint32_t { k_crc_dma_threshold  = 32U; } crc_dma_threshold_t;

Can you clarify where these were supposed to come from?

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -332,17 +344,249 @@ typedef enum : int32_t {
...
+ * // write CRCDIR bytes and read CRCDOR
+ * internal_crc_lock_release();
+ * @endcode
+ *
+ * @see internal_crc_lock_release() Release ownership of the CRC lock
+ * @see rx_crc32_ieee_impl() Uses this lock around CRC hardware transaction
+ *
+ * @since Version 1.2.0
+ */
+static rx_err_t internal_crc_lock_acquire(void)
```
bsikar, line 485: Two things with the spinlock in `internal_crc_lock_acquire()`:

1. `k_crc_lock_sleep_ticks = 0` — in ThreadX, `tx_thread_sleep(0)` returns 
   `TX_CALLER_ERROR` immediately without yielding to any other thread. The 
   comment calls this a "cooperative yield" but it isn't one. Also, the 
   error check right after it:

     if (sleep_status != TX_SUCCESS) { return k_rx_err_invalid_state; }

   ...means the very first time two threads contend for this lock, the 
   second caller gets an immediate error return. That's not a lock, that's 
   a trylock with bad documentation. Change to `k_crc_lock_sleep_ticks = 1U`.

2. The lock uses `tx_interrupt_control(TX_INT_DISABLE)` to achieve atomicity,
   which disables ALL interrupts on the CPU — including motor control ISRs, 
   encoder ISRs, and SPI ISRs — for the duration of the lock check. ThreadX 
   has `tx_mutex_t` for exactly this use case. Can you explain why a full 
   interrupt disable was chosen over a proper mutex here?

e2-studio-star-rx72n-firmware/libs/rx_crc/src/rx_crc32_hw.c
```diff
@@ -332,17 +344,249 @@ typedef enum : int32_t {
...
+ * // CRC critical section: DORCLR, CRCDIR writes, CRCDOR read
+ * internal_crc_lock_release();
+ * @endcode
+ *
+ * @see internal_crc_lock_acquire() Acquire ownership of the CRC lock
+ * @see rx_crc32_ieee_impl() Locking call site around hardware transaction
+ *
+ * @since Version 1.2.0
+ */
+static rx_err_t internal_crc_lock_release(void)
```
bsikar, line 557: In `internal_crc_lock_release()`, on the early error path:

  if (!s_crc_lock) {
      UINT restore_state = tx_interrupt_control(saved);  // restore_state is never read
      return k_rx_err_invalid_state;
  }

`restore_state` is declared and assigned but never used before the return. 
With `-Wall -Wextra -Werror` (which is the project standard per Rule 10), 
this is a build failure. The interrupt restore still needs to happen, but 
the return value should be explicitly discarded:

  (void)tx_interrupt_control(saved);

bsikar (changes requested):
Also claude found these issues:

BUG-4 — Unused variable will break the build under -Werror

In `internal_crc_lock_release()`, on the early error path:

  if (!s_crc_lock) {
      UINT restore_state = tx_interrupt_control(saved);  // restore_state is never read
      return k_rx_err_invalid_state;
  }

`restore_state` is declared and assigned but never used before the return. 
With `-Wall -Wextra -Werror` (which is the project standard per Rule 10), 
this is a build failure. The interrupt restore still needs to happen, but 
the return value should be explicitly discarded:

  (void)tx_interrupt_control(saved);

BUG-5 — NULL dereference risk in rx_dmaca_configure()

After `internal_validate_channel()` passes, the code does:

  ch = dmac_channel(channel);
  ch->dmcnt &= ~(uint8_t)k_dmaca_dte_enable;  // ch not checked before use

`rx_dmaca_start()` (line ~897) checks `ch != NULL` after `dmac_channel()` —
but `rx_dmaca_configure()` doesn't. If `dmac_channel()` ever returns NULL 
for a valid-range channel, this is a NULL dereference. The check is one line, 
and the precedent is right there in the same file. Can you add it here too?

BUG-6 — The re-initialization assert is always false

In `rx_dmaca_init()`:

  if (s_dmaca_initialized) {
      RX_ASSERT(!s_dmaca_initialized, "re-init without reset");
      return k_rx_ok;
  }

Inside the `if (s_dmaca_initialized)` block, `s_dmaca_initialized` is 
guaranteed to be true, so `!s_dmaca_initialized` is always false. The 
assert never fires in any build — debug or release. 

Additionally, a re-initialization returning `k_rx_ok` in release builds 
silently tells the caller "everything is fine" when the docs explicitly 
call this a programming error. This should probably return 
`k_rx_err_invalid_state` instead of success. 

What was the intent here?

BUG-7 — rx_crc32_ieee_impl() exceeds the 60-line function limit

`rx_crc32_ieee_impl()` runs approximately 96 lines in its current form 
with the `#ifdef RX_CRC32_USE_DMA` blocks included. NASA Rule 4 caps 
functions at ~60 lines.

The DMA-dispatch block (the `if (len >= RX_CRC32_DMA_THRESHOLD)` section) 
is a natural extraction point — pull it into an `internal_crc_process_data()` 
helper that takes `data` and `len` and handles the DMA/CPU routing. That 
brings both functions under the limit.

BUG-8 — k_dmaca_num_channels is documented but never defined

Every function in `rx_dmaca.h` documents its channel parameter as 
"Values >= k_dmaca_num_channels result in k_rx_err_nack" — but 
`k_dmaca_num_channels` is never defined anywhere as an enumerator.

The implementation uses `k_dmac_channel_count` (from `rx72n_dmac_regs.h`). 
A caller who reads the header and tries to use `k_dmaca_num_channels` for 
a bounds check gets a compile error.

`dmaca_hw_limit_t` is the right home for this — the enum already covers 
hardware limits:

  typedef enum : uint16_t {
      k_dmaca_max_transfer_count = 1024U,
      k_dmaca_num_channels       = 8U,
  } dmaca_hw_limit_t;

STRUCTURAL — rx_dmaca.c is in the wrong directory

`rx_dmaca.c` lives in `libs/rx_crc/src/` but its header is in 
`libs/rx_hal/inc/rx_dmaca.h`. The DMACA driver has no CRC-specific logic 
in it — it's a general hardware abstraction layer.

Having the implementation inside the CRC library while the header is in 
the HAL is inconsistent and will cause confusion for anyone trying to 
find, extend, or test this module. It should be at 
`libs/rx_hal/src/rx_dmaca.c` alongside the other HAL sources.

STYLE — OR-ing zero "for clarity" is just noise

In `internal_configure_transfer_mode()`:

  /* DMTMD_SZ_BYTE is zero (0U), so this shifts zero into place.
   * Kept explicitly for code clarity... */
  dmtmd |= (k_dmtmd_sz_byte << k_dmtmd_sz_shift);

`k_dmtmd_sz_byte = 0`, so `0 << anything = 0`, so `x |= 0` is a no-op. 
Keeping a no-op "for clarity" actually has the opposite effect — a reader 
now has to verify this isn't a bug before they can move on. Just remove it.

STYLE — Duplicate @retval in rx_dmaca_wait

In the `rx_dmaca_wait()` Doxygen, `k_rx_err_nack` is listed twice:

  @retval k_rx_err_nack    Invalid channel index or channel not configured.
  @retval k_rx_err_nack    validate_channel() failure.

These are the same return code describing the same class of error. 
Consolidate them into one entry.

STYLE — Missing newline at end of rx_dmaca.h and rx_dmaca.c

Both `rx_dmaca.h` and `rx_dmaca.c` are missing a trailing newline 
(the diff shows `\ No newline at end of file`). POSIX requires text 
files to end with a newline, and some compiler toolchains warn on this. 
The pre-commit hook should have caught it.

PR #396, opened 2026-03-03
Jbartz/yml header format checker

jared-bartz-tamu:
created a .yml file that checks files to see if they have the correct copyright, that being * @copyright Copyright (c) 2026 STAR Project

also updated to fix all copyright headers in all .c and .h files.


bsikar:
I will take control over this

PR #447, opened 2026-04-15
New UI

jared-bartz-tamu:
UI is ready to be reviewed. UI elements all look fine and all the buttons seem to work, but testing is needed with gamepad and other things to see how the physical elements react.

## Summary by CodeRabbit

* **New Features**
  * Route-driven navigation with Main Dashboard and ROS Debug views; live SLAM map canvas, trace chart, IMU orientation graphic, and reusable dashboard UI primitives; new operator/controller hooks.

* **UI / Navigation / Style**
  * Dark-first redesign, new local fonts, consolidated design tokens, responsive layouts, updated header/controls and alerts/resume flows.

* **Removed**
  * Legacy tiling/layout system and many legacy dashboard panels/components.

* **Tests**
  * Added tests for primitives, map, traces, hooks, and core views.


star-ui/src/components/StarMapCanvas.tsx
```diff
@@ -0,0 +1,223 @@
...
+      context.beginPath();
+      context.moveTo(0, y);
+      context.lineTo(canvasWidth, y);
+      context.stroke();
+    }
+    context.restore();
+
+    if (pathRef.current.length > 1) {
+      context.save();
+      context.strokeStyle = '#facc15';
```
CesMag: remove magic color

star-ui/src/components/StarMapCanvas.tsx
```diff
@@ -0,0 +1,223 @@
...
+        context.stroke();
+      }
+      context.restore();
+    }
+
+    context.save();
+    context.translate(canvasWidth / 2, canvasHeight / 2);
+    context.rotate(odometry?.thetaRad ?? 0);
+
+    context.fillStyle = '#111827';
```
CesMag: more magic

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
...
+      { label: 'wheel_diameter', value: '0.20 m' },
+      { label: 'imu_sample_rate', value: telemetry?.imu ? '100 Hz' : '--' },
+      { label: 'battery_cells', value: battery?.cells?.validCells ? `${battery.cells.validCells}S` : '--' },
+      { label: 'controller_frequency', value: packets.some((packet) => packet.type === 'controller') ? '50 Hz' : '--' },
+      { label: 'firmware_version', value: systemStatus?.firmwareVersion || '--' },
+    ],
+    [battery, lidarPointCount, odometry, packets, systemStatus, telemetry],
+  );
+
+  const logLines = useMemo<LogLine[]>(() => {
```
CesMag: **Build blocker (TS2345).** The inner `.map(...)` in `packetLogs` produces literals where `tone` is inferred as `string` (not `Tone`), so this whole callback returns `{...}[]` not `LogLine[]`.

Run `npm run build` locally - you'll see:
```
src/App.tsx(1234,39): error TS2345: Type 'string' is not assignable to type 'Tone'.
```

Fix: type the mapped callbacks, e.g. `.map((packet): LogLine => ({ ... }))` on both `packetLogs` and `alertLogs`, or assert `tone` with `as Tone`. This should never have gotten past `tsc -b`.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
...
+      connectionState: state.connectionState,
+      lidarScan: state.lidarScan,
+      odometry: state.odometry,
+      systemStatus: state.systemStatus,
+      telemetry: state.telemetry,
+    })),
+  );
+
+  const packets = usePacketFeed();
+  const now = Date.now();
```
CesMag: **Lint error: `react-hooks/purity`.** `Date.now()` is impure and cannot be called during render in React 19. This will produce unstable results as the component re-renders.

Either compute `now` inside the `useMemo`/`useEffect` that actually needs it, or stash it in a `ref`/`state` updated by an effect. Don't call `Date.now()` at the top level of the component body.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
...
+            ? 'accent'
+            : 'good',
+      source: packet.direction === 'tx' ? 'TX' : 'RX',
+      message: packet.preview || packet.type,
+      emphasis: packet.type,
+    }));
+
+    const alertLogs = alerts.slice(0, 10).map((alert, index) => ({
+      key: `alert-${alert.code}-${alert.timestampUs}-${index}`,
+      tsMs: Date.now() - index * 750,
```
CesMag: **Lint error: `react-hooks/purity`.** Same issue as line 1091 - `Date.now()` inside a `useMemo` callback that runs during render. The fake `tsMs: Date.now() - index * 750` is also suspicious; why are alert timestamps being synthesized from `Date.now()` instead of using the real `timestampUs` from the alert? If you need a deterministic ordering, use the actual timestamp the alert carries.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
...
+  const controllerBridgeEnabled = uiMode === 'manual' && !eStopActive;
+  const gamepadState = useControllerBridge(sendControllerState, controllerBridgeEnabled);
+
+  useEffect(() => {
+    if (!systemStatus) {
+      return;
+    }
+
+    if (systemStatus.mode === RobotMode.MANUAL) {
+      setUiMode('manual');
```
CesMag: **Lint error: `react-hooks/set-state-in-effect`.** Synchronous `setState` inside a `useEffect` body triggers cascading renders. Rewrite this so the mode transition is computed and set in one place, or move it into an event handler instead of an effect that reacts to `systemStatus`. Also fires at line 525 and 531 for similar reasons - consolidate the `uiMode`/`autonomyRunning` state machine.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
-// LAYOUT ENGINE: Tiling WM (react-grid-layout) with view-based layout system
-import { useEffect, useRef, useState, type ReactNode } from 'react';
+import { useEffect, useMemo, useRef, useState, type MouseEvent as ReactMouseEvent, type ReactNode } from 'react';
```
CesMag: **Dead code left behind.** This PR imports only `DebugTraceChart`, `StarMapCanvas`, `useGamepad`, `useSTARConnection`, proto, and store modules. The following 20 components in `src/components/` are no longer imported anywhere in `src/`:

AlertsPanel, BatteryPanel, CameraFeed, ControllerView, DiagLogPanel, FirmwarePanel, GpsPanel, ImuPanel, LidarPanel, MappingPanel, MotorPanel, Nav2GoalPanel, OdometryPanel, PacketAnalyzer, PidTuningPanel, StatusBar, SystemHealthPanel, TeleopPanel, TimeSeriesPanel, TransportDiagPanel.

Per `CLAUDE.md` (Backward Compatibility Policy): delete orphaned code in the same PR that replaces it. Do not leave these sitting as future archaeology.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
...
+  lidar: '/scan',
+  odometry: '/odom',
+  controller: '/cmd_vel',
+  telemetry: '/imu/data',
+  battery: '/battery_state',
+  system: '/tf',
+  alert: '/diagnostics',
+};
+
+const staticTopicRows = [
```
CesMag: `staticTopicRows` hardcodes rates like `'10 Hz'`, `'50 Hz'`, `'100 Hz'`, `'1 Hz'`, `'200 Hz'`. A "ROS2 Debug Console" that shows **static strings** instead of the real topic rates is actively misleading - an operator will read these numbers and believe them. Either compute from live packet timestamps or drop the column.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,1441 @@
...
-    firmware: <FirmwarePanel />,
-    nav2: <Nav2GoalPanel />,
-    diaglog: <DiagLogPanel />,
-  };
+type AppRoute = '/' | '/ros';
+type UiMode = 'autonomous' | 'manual';
+type Tone = 'neutral' | 'good' | 'warn' | 'danger' | 'accent';
+
+const controllerSendIntervalMs = 20;
+const packetPollIntervalMs = 1000;
```
CesMag: `coverageStepPercent = 1.5` plus the interval at `setCoveragePercent(...)` later in the file is a fake progress bar. This is demo-ware. It will tick up 1.5% every ~450ms regardless of what the robot is actually doing. Either remove it or explicitly gate behind a `VITE_DEMO_MODE` flag with a visible banner - do not ship simulated autonomy telemetry into a UI people will use to actually operate the robot.

CesMag (changes requested):
This PR does not build and does not lint. It cannot be reviewed as "ready" in this state.

## Hard blockers

1. `npm run build` fails. TypeScript error TS2345 in `App.tsx:1234` - `logLines` inner map literals have `tone`/`source` inferred as `string`, so the array is not assignable to `LogLine[]`. This is not a subtle issue; it is the first thing `tsc -b` reports. The PR could not have been built locally before being marked ready for review.
2. `npm run lint` produces 5 errors (React 19 purity and hooks rules). Details inline.
3. The 20 existing panel components (AlertsPanel, BatteryPanel, CameraFeed, ControllerView, DiagLogPanel, FirmwarePanel, GpsPanel, ImuPanel, LidarPanel, MappingPanel, MotorPanel, Nav2GoalPanel, OdometryPanel, PacketAnalyzer, PidTuningPanel, StatusBar, SystemHealthPanel, TeleopPanel, TimeSeriesPanel, TransportDiagPanel) are not imported anywhere after this PR. Per the project's zero-backward-compatibility policy in CLAUDE.md, this is not acceptable - either delete them in this PR or restore their usage. Leaving ~20 orphaned top-level components is not "new UI," it is half a rewrite.

## Quality issues

- `App.tsx` is ~1440 lines. It contains the entire application, route handling, all views, all interfaces (`LogLine`, `EStopResumeSnapshot`, `TraceHistory`, `HeaderLink`, `Glyph`), and multiple hooks. Split it into `src/views/`, `src/components/`, and `src/hooks/`. No one wants to review or maintain a 1.4k-line `App.tsx`.
- `staticTopicRows` and `parameterCards` ship hardcoded fake telemetry (`'10 Hz'`, `'0.05 m/cell'`, `'0.20 m'`, `'100 Hz'`). Presenting simulated values through a "debug console" UI is actively misleading on the bench. Wire them to real state or remove them.
- `setCoveragePercent` with a fixed step (`coverageStepPercent = 1.5`) is demo-ware masquerading as autonomy progress. Gate it behind an explicit demo flag or delete it.
- Hand-rolled router against `window.location.pathname` + `popstate` for two routes - acceptable, but at this component size a state-driven view switch or `react-router` would be clearer and avoid full-reload surprises.

## Testing

- Zero tests for the new `App.tsx`, `StarMapCanvas`, or `DebugTraceChart`. The existing `useGamepad.test.ts` is the only test file. Add at least render smoke tests for each new component.
- PR description admits gamepad testing has not been done, yet the UI is marked ready for review. Manual test the gamepad + e-stop + manual/autonomous mode path before re-requesting review.

## Required before I'll look at this again

1. `npm run build` passes.
2. `npm run lint` passes (fix all 5 errors, don't disable rules).
3. Old panel components deleted (not commented, not moved - deleted).
4. Fake/simulated telemetry gated behind an explicit demo flag or removed.
5. At least one render test per new component.
6. Gamepad + e-stop manually validated against gateway.

See inline comments for the specific lines.

Scicada46:
Screenshot 2026-04-17 at 4 00 00 AM
Screenshot 2026-04-17 at 4 00 10 AM
Some screenshots of ui running

bsikar (commented):
Reviewing what's actually been addressed since the last CodeRabbit pass on 2026-04-17.

**TL;DR:** of the 36 new CodeRabbit comments from that review, 0 appear to have been touched. The 2026-04-15 CodeRabbit batch and most of the earlier review feedback were addressed (monolithic `App.tsx` split, 20 orphan panels deleted, demo mode gated behind `VITE_DEMO_MODE`) - but the follow-up round was not. Inline comments below cover the high-impact ones plus a few things unique to this pass.

ASCII-only is a hard project rule (`CLAUDE.md`) - the `°` and `·` literals below will fail the pre-commit hook as soon as anyone else touches these files.

star-ui/src/App.tsx
```diff
@@ -1,202 +1,25 @@
...
-    gps: <GpsPanel />,
-    timeseries: <TimeSeriesPanel />,
-    transport: <TransportDiagPanel />,
-    pid: <PidTuningPanel />,
-    firmware: <FirmwarePanel />,
-    nav2: <Nav2GoalPanel />,
-    diaglog: <DiagLogPanel />,
-  };
-}
+const wsPort = Number(import.meta.env.VITE_WS_PORT) || 8080;
```
bsikar: `Number(import.meta.env.VITE_WS_PORT) || 8080` silently swallows `0` and `NaN` and falls back to 8080. If someone sets `VITE_WS_PORT=0` (or mistypes it as a non-numeric string) they get 8080 instead of an error. Parse explicitly and validate:

```ts
const parsed = Number.parseInt(import.meta.env.VITE_WS_PORT ?? '', 10);
const wsPort = Number.isFinite(parsed) && parsed > 0 && parsed < 65536 ? parsed : 8080;
```

star-ui/src/lib/dashboard.ts
```diff
@@ -0,0 +1,211 @@
...
+  return value.toFixed(digits);
+}
+
+export function formatHeading(radians: number | undefined): string {
+  if (radians == null || Number.isNaN(radians)) {
+    return '--';
+  }
+
+  const degrees = ((radians * 180) / Math.PI + 360) % 360;
+  return `${degrees.toFixed(1)}°`;
```
bsikar: `` `${degrees.toFixed(1)}°` `` - non-ASCII `°` (U+00B0). This violates the project's ASCII-only source policy in `CLAUDE.md` and will be rejected by `scripts/git/pre-commit`. Replace with `' deg'` or unicode-escape it: `` `${degrees.toFixed(1)}\u00B0` ``.

star-ui/src/views/MainDashboardView.tsx
```diff
@@ -0,0 +1,581 @@
...
+                    <div className="subpanel-card__content subpanel-card__content--split">
+                      <ImuOrientationGraphic
+                        pitchDeg={imu ? (imu.pitchRad * 180) / Math.PI : 0}
+                        rollDeg={imu ? (imu.rollRad * 180) / Math.PI : 0}
+                      />
+
+                      <div className="orientation-values">
+                        <MetricTile label="Roll" tone="accent" value={`${formatNumber(imu ? (imu.rollRad * 180) / Math.PI : undefined, 1)}°`} />
+                        <MetricTile label="Pitch" tone="accent" value={`${formatNumber(imu ? (imu.pitchRad * 180) / Math.PI : undefined, 1)}°`} />
+                        <MetricTile label="Yaw" tone="accent" value={`${formatNumber(headingRadians != null ? (headingRadians * 180) / Math.PI : undefined, 1)}°`} />
```
bsikar: Three non-ASCII `°` characters here (lines 364, 365, 366 for Roll/Pitch/Yaw `MetricTile` values). Same ASCII-only policy violation as `dashboard.ts:84`. Either use `' deg'` or `\u00B0`.

star-ui/src/views/RosDebugView.tsx
```diff
@@ -0,0 +1,307 @@
...
+            <div className="panel-card__body">
+              {demoModeEnabled ? (
+                <div className="metrics-grid metrics-grid--three">
+                  {demoParameterCards.map((parameter) => (
+                    <MetricTile key={parameter.label} label={parameter.label} tone="warn" value={parameter.value} />
+                  ))}
+                </div>
+              ) : (
+                <div className="empty-state">
+                  Live parameter values are unavailable from the current gateway. Firmware {systemStatus?.firmwareVersion || '--'} · Uptime{' '}
```
bsikar: Non-ASCII middle-dot `·` (U+00B7) in the fallback copy. ASCII-only policy - use `-` or ` | `.

star-ui/src/views/MainDashboardView.tsx
```diff
@@ -0,0 +1,581 @@
...
+                    <div className="subpanel-card__header">
+                      <h3>Encoders &amp; Odometry</h3>
+                      <p>Wheel RPM estimates and global pose tracking.</p>
+                    </div>
+                    <div className="subpanel-card__content">
+                      <div className="encoder-odometry-layout">
+                        <div className="sensor-metrics sensor-metrics--encoders">
+                          {['FL', 'FR', 'RL', 'RR'].map((label, index) => {
+                            const motor = motorValues[index];
+                            const rpm = motor ? (motor.velocityMps / 0.2) * 60 : undefined;
```
bsikar: **Wrong formula.** `(motor.velocityMps / 0.2) * 60` treats the wheel diameter `0.2 m` as its circumference. For a real wheel of diameter `d`, RPM is `velocityMps * 60 / (pi * d)`. As written, this overreads by a factor of `pi` - a motor spinning at 100 RPM will display as ~314. Either fix the math or name the constants (`k_wheel_circumference_m`, `k_sec_per_min`) so this kind of mistake is obvious on review.

star-ui/src/views/RosDebugView.tsx
```diff
@@ -0,0 +1,307 @@
...
+  { name: '/nav2_controller', description: 'Navigation2 path planning', tone: 'good' as const },
+  { name: '/robot_state_publisher', description: 'Robot kinematic model', tone: 'good' as const },
+];
+
+const demoParameterCards = [
+  { label: 'scan_frequency', value: '10 Hz' },
+  { label: 'map_resolution', value: '0.05 m/cell' },
+  { label: 'wheel_diameter', value: '0.20 m' },
+  { label: 'controller_frequency', value: '50 Hz' },
+];
```
bsikar: This was CesMag's comment #6 on the previous review and it was not addressed. `staticTopicRows` still hardcodes `'10 Hz' / 50 Hz / 100 Hz / 200 Hz` in a UI labeled **ROS2 Debug Console**. An operator reading this will believe those are live rates. Compute from packet timestamps in `usePacketFeed` or delete the rate column - do not ship static lies in a debug view.

star-ui/src/components/StarMapCanvas.tsx
```diff
@@ -0,0 +1,223 @@
...
+    context.clearRect(0, 0, canvasWidth, canvasHeight);
+
+    context.fillStyle = '#0d1520';
+    context.fillRect(0, 0, canvasWidth, canvasHeight);
+
+    context.save();
+    context.strokeStyle = 'rgba(112, 145, 183, 0.18)';
+    context.lineWidth = 1;
+    for (let offset = -20; offset <= 20; offset += gridSpacingMeters) {
+      const x = canvasWidth / 2 + offset * pixelsPerMeter - ((origin.x * pixelsPerMeter) % (gridSpacingMeters * pixelsPerMeter));
```
bsikar: Modulo on a possibly-negative origin: `(offset % gridSize)` returns negative values for negative `offset` in JS, which makes the grid visibly drift by `gridSize` when odometry crosses 0. Use a non-negative mod helper:

```ts
const mod = (a: number, n: number) => ((a % n) + n) % n;
```

Applies on line 99 too.

star-ui/src/components/StarMapCanvas.tsx
```diff
@@ -0,0 +1,223 @@
...
+      if (projected.x < -8 || projected.x > canvasWidth + 8 || projected.y < -8 || projected.y > canvasHeight + 8) {
+        return;
+      }
+      context.beginPath();
+      context.arc(projected.x, projected.y, 1.25, 0, Math.PI * 2);
+      context.fill();
+    });
+    context.restore();
+
+    if (lidarScan) {
```
bsikar: `if (lidarScan)` renders beams even when `odometry` is null, using `odometry?.thetaRad ?? 0` as the robot heading. That produces 'phantom' beams pointing in a direction the robot isn't actually facing during the gap between the first lidar packet and the first odometry packet. Gate on both: `if (lidarScan && odometry)`.

star-ui/src/hooks/useControllerBridge.ts
```diff
@@ -0,0 +1,44 @@
...
+      sendRef.current({
+        linearVel: 0,
+        angularVel: 0,
+        timestamp: String(Date.now()),
+        debug: false,
+      });
+      return;
+    }
+
+    const interval = window.setInterval(() => {
```
bsikar: No immediate tick on enable - the first controller state frame is delayed by `controllerSendIntervalMs` (50ms at 20 Hz, 20ms at 50 Hz). At 20 Hz that's a 50ms latency spike every time the user switches to manual. Call `sendRef.current(...)` once before `setInterval` is scheduled, or use `setTimeout(..., 0)` to fire on the same tick.

star-ui/src/hooks/useOperatorControls.ts
```diff
@@ -0,0 +1,175 @@
...
+        setAutonomyRequested(snapshot.wasAutonomyRequested);
+      }
+
+      if (!released) {
+        store.addAlert({
+          code: 'ESTOP_RESUME_LOCAL',
+          level: AlertLevel.WARN,
+          message: 'Resume applied locally; backend release confirmation was not received.',
+          source: 'ui',
+          timestampUs: String(Date.now() * 1000),
```
bsikar: `timestampUs: String(Date.now() * 1000)` - magic `1000` appears here and on line 137. Given this project's explicit 'no magic numbers' rule in `CLAUDE.md`, extract `const msToUs = 1000;` (or import the one already defined in `useSTARConnection.ts:7`) and reuse it. It's already a named constant in one file and a magic number in the next.

star-ui/src/hooks/useOperatorControls.ts
```diff
@@ -0,0 +1,175 @@
...
+          setAutonomyRequested(false);
+        }
+        return nextCoverage;
+      });
+    }, coverageTickIntervalMs);
+
+    return () => window.clearInterval(interval);
+  }, [autonomyRunning]);
+
+  async function handleEStopToggle(): Promise<void> {
```
bsikar, line 98: The `await sendEStopRelease()` path calls `setState` (or triggers parent handlers that do) after the await, with no mount guard. If the component unmounts during the fetch fallback, this produces a React warning at best and a stale-closure bug at worst. Wrap with a `mountedRef` check or an `AbortController`.

star-ui/src/hooks/usePacketFeed.ts
```diff
@@ -0,0 +1,28 @@
...
+
+interface PacketFeedState {
+  packets: PacketRecord[];
+  sampledAtMs: number;
+}
+
+export function usePacketFeed(): PacketFeedState {
+  const [feed, setFeed] = useState<PacketFeedState>({
+    packets: getRecentPackets(),
+    sampledAtMs: 0,
```
bsikar: `sampledAtMs: 0` as the initial sentinel is fine until consumers start formatting it ('last update: 1970-01-01'). Use `null` with a `number | null` type so downstream code is forced to handle 'never sampled' explicitly.

star-ui/src/hooks/useTraceHistory.ts
```diff
@@ -0,0 +1,39 @@
...
+  velocity: number;
+}
+
+export function useTraceHistory(samples: TraceSamples): TraceHistory {
+  const latestTraceSamplesRef = useRef(samples);
+  const [traceHistory, setTraceHistory] = useState<TraceHistory>(() => ({
+    cpu: Array.from({ length: traceHistoryLength }, () => samples.cpu),
+    lidar: Array.from({ length: traceHistoryLength }, () => samples.lidar),
+    temperature: Array.from({ length: traceHistoryLength }, () => samples.temperature),
+    velocity: Array.from({ length: traceHistoryLength }, () => samples.velocity),
```
bsikar: Seeds the buffer with 28 zero-value samples on mount. Result: the trace chart shows a fake flatline at 0 for the first ~28 seconds before real telemetry arrives - indistinguishable from 'robot reporting all-zero telemetry'. Seed with `NaN` (and have the chart skip NaN) or start empty and grow as samples arrive.

star-ui/src/hooks/useTraceHistory.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    const interval = window.setInterval(() => {
+      setTraceHistory((currentHistory) => ({
+        cpu: pushTraceSample(currentHistory.cpu, latestTraceSamplesRef.current.cpu),
+        lidar: pushTraceSample(currentHistory.lidar, latestTraceSamplesRef.current.lidar),
+        temperature: pushTraceSample(currentHistory.temperature, latestTraceSamplesRef.current.temperature),
+        velocity: pushTraceSample(currentHistory.velocity, latestTraceSamplesRef.current.velocity),
+      }));
+    }, 1000);
```
bsikar: Hardcoded `1000` - this is the trace sample interval in ms. Import or define `traceSampleIntervalMs` in `lib/dashboard.ts` next to the other dashboard constants. Same magic-number rule as `useOperatorControls.ts`.

star-ui/src/hooks/useAppRoute.ts
```diff
@@ -0,0 +1,25 @@
...
+  const [route, setRoute] = useState<AppRoute>(() => normalizeRoute(window.location.pathname));
+
+  useEffect(() => {
+    const handlePopState = () => setRoute(normalizeRoute(window.location.pathname));
+    window.addEventListener('popstate', handlePopState);
+    return () => window.removeEventListener('popstate', handlePopState);
+  }, []);
+
+  function navigate(nextRoute: AppRoute): void {
+    if (normalizeRoute(window.location.pathname) !== nextRoute) {
```
bsikar: `navigate` short-circuits when the target matches `normalizeRoute(window.location.pathname)`. If the actual `pathname` is e.g. `/ros/` or `/ROS`, `normalizeRoute` collapses it to `/ros`, so `navigate('/ros')` becomes a no-op and the address bar keeps the un-normalized URL. Compare against the internal `route` state (the source of truth) instead of re-normalizing the location.

star-ui/src/views/MainDashboardView.tsx
```diff
@@ -0,0 +1,581 @@
...
+      return;
+    }
+
+    function handleOutsideClick(event: globalThis.MouseEvent): void {
+      if (alertsRef.current && !alertsRef.current.contains(event.target as Node)) {
+        setAlertsOpen(false);
+      }
+    }
+
+    document.addEventListener('mousedown', handleOutsideClick);
```
bsikar, line 193: Only `mousedown` closes the alerts popover. On touch devices this leaves the popover stuck open after a tap outside; keyboard users also can't dismiss it with Escape. Add `touchstart` and a `keydown` (Escape) handler.

star-ui/src/views/MainDashboardView.tsx
```diff
@@ -0,0 +1,581 @@
...
+            <HeaderLink href="/ros" onNavigate={navigate} label="Debug" accent="accent" glyph="RX" />
+            <Chip tone={connectionState === 'connected' ? 'good' : connectionState === 'disconnected' ? 'danger' : 'warn'}>
+              {connectionState === 'connected' ? 'Connected' : connectionState === 'reconnecting' ? 'Reconnecting' : connectionState}
+            </Chip>
+            {uiMode === 'manual' ? (
+              <Chip tone={gamepadState.connected ? 'good' : 'warn'}>{gamepadState.connected ? 'Gamepad' : 'No Gamepad'}</Chip>
+            ) : null}
+
+            <div className="alerts-anchor" ref={alertsRef}>
+              <AlertsButton count={alertPreview.length} onClick={handleAlertsToggle} />
```
bsikar: `<AlertsButton count={alertPreview.length} />` - `alertPreview` is capped (looks like 6 from the slice upstream). The badge therefore tops out at 6 even when there are 200 alerts in the store. Pass `alerts.length` instead, or cap visually with `99+`.

star-ui/src/views/RosDebugView.tsx
```diff
@@ -0,0 +1,307 @@
...
+                <p>Project Star - System Diagnostics</p>
+              </div>
+            </div>
+          </div>
+
+          <div className="screen-header__actions">
+            {demoModeEnabled ? <Chip tone="warn">Demo Mode</Chip> : null}
+            <Chip tone={connectionState === 'connected' ? 'good' : connectionState === 'disconnected' ? 'danger' : 'warn'}>
+              {connectionState === 'connected' ? 'Connected' : connectionState === 'reconnecting' ? 'Reconnecting' : connectionState}
+            </Chip>
```
bsikar: The `connectionState === 'connecting'` branch falls through and renders the raw `{connectionState}` token (`'connecting'`) as label text. All four states should have a human-facing label or this should use a lookup table; relying on the enum value to be presentable is a latent bug.

star-ui/src/lib/dashboard.ts
```diff
@@ -0,0 +1,211 @@
...
+export const controllerSendIntervalMs = 20;
+export const packetPollIntervalMs = 1000;
+export const recentPacketWindowMs = 8_000;
+export const coverageTickIntervalMs = 450;
+export const coverageStepPercent = 1.5;
+export const traceHistoryLength = 28;
+
+const validAlertTimestampFloorMs = Date.parse('2020-01-01T00:00:00Z');
+
+export const demoModeEnabled = import.meta.env.VITE_DEMO_MODE === 'true';
```
bsikar, line 56: `import.meta.env.VITE_DEMO_MODE` is untyped - there's no `vite-env.d.ts` declaring `VITE_DEMO_MODE: string` on `ImportMetaEnv`, so autocomplete/typecheck treat it as `any`. Add `src/vite-env.d.ts` with an `ImportMetaEnv` interface extension covering `VITE_WS_PORT` and `VITE_DEMO_MODE`.

star-ui/src/test/setup.ts
```diff
@@ -1 +1,4 @@
 import '@testing-library/jest-dom';
+import { vi } from 'vitest';
+
+HTMLCanvasElement.prototype.getContext = vi.fn(() => null);
```
bsikar: `HTMLCanvasElement.prototype.getContext = () => null` - any component that does `const ctx = canvas.getContext('2d'); ctx.fillRect(...)` under test will throw a NullPointerException-equivalent. `StarMapCanvas` only works today because the test happens to exit early. Provide a stub context object (`{ fillRect: vi.fn(), beginPath: vi.fn(), ... }`) so tests exercise the real draw paths.

star-ui/src/views/MainDashboardView.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { useDashboardStore } from '../store/dashboardStore';
+import { MainDashboardView } from './MainDashboardView';
+
+vi.mock('../hooks/useControllerBridge', () => ({
+  useControllerBridge: () => ({
+    connected: false,
+    linearVel: 0,
+    angularVel: 0,
+  }),
```
bsikar: The `useControllerBridge` mock returns `{ connected, linearVel, angularVel }` but the real hook returns a full `GamepadState`. Mocks drifting from real shape hides type regressions. Either `vi.mocked<typeof useControllerBridge>` it or mock the whole `useGamepad` surface.

star-ui/src/views/RosDebugView.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      connectionState: 'connected',
+      lidarScan: null,
+      odometry: null,
+      systemStatus: null,
+      telemetry: null,
+    });
+  });
+
+  it('renders the debug console shell', () => {
+    render(<RosDebugView navigate={vi.fn()} />);
```
bsikar, line 43: `navigate={vi.fn()}` is passed but never asserted - the test doesn't verify that clicking 'Return to Main' actually calls `navigate('/')`. Add the assertion or drop the prop.

star-ui/src/components/DebugTraceChart.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+        formatValue={(value) => `${value.toFixed(1)} m/s`}
+        samples={[0.2, 0.4, 0.8]}
+        title="Velocity"
+      />,
+    );
+
+    expect(screen.getByText('Velocity')).toBeInTheDocument();
+    expect(screen.getByText('0.8 m/s')).toBeInTheDocument();
+    expect(screen.getByText('3 pts')).toBeInTheDocument();
+  });
```
bsikar, line 19: Only a single happy-path snapshot. No coverage for empty samples, single-sample, or NaN-containing samples - all of which are reachable through `useTraceHistory`. Add at least an empty-samples test.

star-ui/src/components/StarMapCanvas.test.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { describe, expect, it } from 'vitest';
+import { StarMapCanvas } from './StarMapCanvas';
+
+describe('StarMapCanvas', () => {
+  it('renders the empty-state overlay when odometry is unavailable', () => {
+    const { container } = render(<StarMapCanvas lidarScan={null} odometry={null} />);
+
+    expect(screen.getByText('Awaiting live odometry feed...')).toBeInTheDocument();
+    expect(container.querySelector('canvas')).toBeInTheDocument();
+  });
```
bsikar, line 12: Happy-path-only: the test renders with null `odometry` and `lidarScan`. No test exercises the draw path with non-null data - which is exactly where the phantom-beam and negative-origin-modulo bugs live.

star-ui/src/components/dashboard/ImuOrientationGraphic.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface ImuOrientationGraphicProps {
+  pitchDeg: number;
+  rollDeg: number;
+}
+
+export function ImuOrientationGraphic({ pitchDeg, rollDeg }: ImuOrientationGraphicProps) {
+  const visualRoll = clamp(rollDeg * 0.65, -22, 22);
+  const visualPitch = clamp(pitchDeg * 0.42, -7, 7);
+  const deckY = 50 + visualPitch;
+  const arrowY = -30 - visualPitch * 0.35;
```
bsikar: Magic numbers throughout the SVG geometry: `0.65`, `0.42`, `22`, `7`, `50`, `-30`, `0.35`. At minimum give these names (`k_horizon_y_ratio`, `k_roll_indicator_radius`, etc.) so the next person adjusting the graphic doesn't have to reverse-engineer what each literal means. Same `CLAUDE.md` no-magic-numbers rule.

star-ui/src/components/dashboard/DashboardPrimitives.tsx
```diff
@@ -0,0 +1,111 @@
...
+    onNavigate(href);
+  }
+
+  return (
+    <a className="header-link" href={href} onClick={handleClick}>
+      <Glyph accent={accent} label={glyph} />
+      <span>{label}</span>
+    </a>
+  );
+}
```
bsikar, line 92: `HeaderLink` calls `e.preventDefault()` unconditionally. That breaks:
- cmd/ctrl-click (open in new tab)
- middle-click (open in new tab)
- shift-click (open in new window)

Guard with `if (e.metaKey || e.ctrlKey || e.shiftKey || e.button === 1) return;` before `preventDefault`.

star-ui/tsconfig.node.json
```diff
@@ -22,5 +22,5 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true
   },
-  "include": ["vite.config.ts"]
+  "include": ["vitest.config.ts"]
```
bsikar: `include: ["vitest.config.ts"]` but `vite.config.ts` is also a Node-context file and is currently being typechecked under the app tsconfig (where `@types/node` isn't in scope). Add `vite.config.ts` to the `include` array.

